PR #2, opened 2015-12-20
BinarySize for size of encoded objects

ebuchman:
Motivated by needing to length prefix messages over the network. But maybe nice to have regardless.


ebuchman:
Closing until someone really needs it

PR #13, opened 2017-04-03
Make time parsing more flexible

ethanfrey:
Handle variable number of decimal places for the second when parsing json strings into time.Time.

Resolves https://github.com/tendermint/tendermint/issues/440

### DIFF
--- a/reflect.go
+++ b/reflect.go
@@ -88,7 +88,7 @@ var (
 )
 
 const (
-	iso8601 = "2006-01-02T15:04:05.000Z" // forced microseconds
+	RFC3339Millis = "2006-01-02T15:04:05.000Z" // forced microseconds
 )
 
 // NOTE: do not access typeInfos directly, but call GetTypeInfo()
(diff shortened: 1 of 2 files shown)


melekes (approved):

melekes:
May I suggest another name for this:

```
const (
	iso8601 = "2006-01-02T15:04:05.000Z" // forced microseconds
)
```

**`iso8601millis`**

https://hackage.haskell.org/package/iso8601-time-0.1.4/docs/Data-Time-ISO8601.html

iso8601 does not specify micro or nanoseconds https://en.wikipedia.org/wiki/ISO_8601

ethanfrey:
Please do.  This is also inaccurate, as 2006-01-03T1543,5 is valid iso8601(!) 
https://en.wikipedia.org/wiki/ISO_8601#Times

meaning 15:43:30 (half a minute).  comma or decimal point can be used as a fractional separator in any location. the [golang time lib](https://golang.org/pkg/time/#pkg-constants) refers to RFC3339.

But yes, why not RFC3339Millis?  In comparision with the RFC3339Nanos I use to parse

melekes:
Ah, yes! `RFC3339Millis` would be even better

reflect.go
```diff
@@ -88,7 +88,7 @@ var (
 )
 
 const (
-	iso8601 = "2006-01-02T15:04:05.000Z" // forced microseconds
+	RFC3339Milli = "2006-01-02T15:04:05.000Z" // forced microseconds
```
melekes: RFC3339Millis :) forgot s at the end

ethanfrey: Not any more... thanks!

Issue #14, opened 2017-04-03
go-wire ReadBinaryBytes can hang sometime


ethanfrey:
I have no reproduceable case yet, but I have observed 2 or 3 times now that when running test cases that mutate bytes to generate many thousand examples of mutated data, that occassionally the test will hang.  I am pretty sure it is due to the ReadBinaryBytes method, as that was the only code in common between the two test cases (unless somehow MutateByteSlice causes an inifite loop).

So, keep your eyes open...

ethanfrey:
Found the issue and fixed it in PR #15  Waiting for review...

ethanfrey:
Already merged.

Issue #12, opened 2017-03-29
Panic when feeding invalid data


ethanfrey:
In my tests, I marshal a valid proof to bytes, then fuzz it with MutateBytes, and get quite a few panics (which I had to catch).  It would be nice to make go-wire a bit more robust:

Seems if no limit is passed in (ReadBinaryBytes), this length can overflow.
https://github.com/tendermint/go-wire/blob/master/byteslice.go#L28

```
runtime error: makeslice: len out of range
github.com/tendermint/light-client/proofs.AppProver.Unmarshal.func1
    /Users/ethan/golang/src/github.com/tendermint/light-client/proofs/app.go:64
runtime.call32
    /usr/local/go/src/runtime/asm_amd64.s:479
runtime.gopanic
    /usr/local/go/src/runtime/panic.go:458
runtime.makeslice
    /usr/local/go/src/runtime/slice.go:49
github.com/tendermint/light-client/vendor/github.com/tendermint/go-wire.ReadByteSlice
    /Users/ethan/golang/src/github.com/tendermint/light-client/vendor/github.com/tendermint/go-wire/byteslice.go:28
github.com/tendermint/light-client/vendor/github.com/tendermint/go-wire.readReflectBinary
    /Users/ethan/golang/src/github.com/tendermint/light-client/vendor/github.com/tendermint/go-wire/reflect.go:277
github.com/tendermint/light-client/vendor/github.com/tendermint/go-wire.readReflectBinary
    /Users/ethan/golang/src/github.com/tendermint/light-client/vendor/github.com/tendermint/go-wire/reflect.go:317
github.com/tendermint/light-client/vendor/github.com/tendermint/go-wire.ReadBinaryPtr
    /Users/ethan/golang/src/github.com/tendermint/light-client/vendor/github.com/tendermint/go-wire/wire.go:48
github.com/tendermint/light-client/vendor/github.com/tendermint/go-wire.ReadBinaryBytes
    /Users/ethan/golang/src/github.com/tendermint/light-client/vendor/github.com/tendermint/go-wire/util.go:25
github.com/tendermint/light-client/proofs.AppProver.Unmarshal
    /Users/ethan/golang/src/github.com/tendermint/light-client/proofs/app.go:72
github.com/tendermint/light-client/proofs.(*AppProver).Unmarshal
```

For some strange reason the same fuzz code doesn't cause a panic with this struct:
```Go
type TxProof struct {
	Height uint64
	Index  int
	Txs    types.Txs
}
[...]

PR #15, opened 2017-04-03
Stop overflows in ReadBinaryBytes

ethanfrey:
This addresses #14 and #12 simply by adding a hard limit to the parsed size when we know it in advance.

Maybe this is too hard a limit, but I am sure we can figure out a safe number based on the size of the input.  So far the tendermint tests all run through with these values.


ethanfrey:
@jaekwon I changed `len(d)+1` to `len(d)`.  Do you give it approval now?

Issue #32, opened 2017-09-19
doc.go, provide examples and high level description of go-wire and how to use it


odeke-em:
I realize that go-wire is a low-level codec, comparable with https://golang.org/pkg/encoding/gob/
and this of course requires one to register each type and set its identifying ID or type byte.

However, the package is quite undocumented and requires guessing around in order to use it. We could use a doc.go file at the top level and perhaps some examples.

It took me having to spend hours going through all the pieces in PR https://github.com/tendermint/tendermint/pull/648 and also lost time trying to figure out what was wrong with https://github.com/tendermint/tendermint/issues/652#issuecomment-329334819.

I personally don't think that I am a Go newbie, I am familiar with codecs but even that it has taken me a very long time to stitch things up with go-wire, so I can only empathize with the typical beginner, moreover dealing with advanced concepts such as the blockchain to top it all off.

This is a rant, it's just me trying to articulate why a doc.go file and examples are important to ease with development and enable other developers. I believe we could recoup lost time and also reduce the bus factor with the updates requested.

PR #33, opened 2017-09-20
doc.go, examples_test.go: improve DevExp with doc and examples

odeke-em:
Updates #32 

* Added an overview doc.go file
* Added end to end examples that are runnable and copy+pastable
for RegisterInterface, WriteBinary, ReadBinary

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -1,0 +1,69 @@
+/*
+go-wire is our custom codec package for serializing and deserializing
+data and structures as binary and JSON blobs.
+
+In order to get started with go-wire we need to:
+
+1) Choose the receiving structure for deserializing.
+It MUST be an interface{} and during registration it MUST be
+wrapped as a struct for example
+
+    struct { Receiver }{}
+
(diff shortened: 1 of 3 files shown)


odeke-em:
Exhibit
---
screen shot 2017-09-19 at 6 05 39 pm

screen shot 2017-09-19 at 6 05 51 pm

zramsay (approved):

melekes (approved):

ethanfrey (commented):
This is a very nice write up, but I think the API it documents needs to rewritten (see #23)

I would at least use `BinaryBytes` adn `ReadBinaryBytes` as the typical entry points instead of `WriteBinary` and `ReadBinary`, as they are simpler to use.

But the reason you cannot find those among the 67 other functions is why I think the whole API needs a cleanup.

In short, I love the docs, I don't like the API it documents.

ebuchman:
ill merge this but we should have the API redesign discussion.

PR #35, opened 2017-10-05
failing test for backwards compatibility with omitempty

ebuchman:
Failing test illustrating unexpected behaviour with `omitempty` and binary encoding

### DIFF
--- a/reflect_test.go
+++ b/reflect_test.go
@@ -803,3 +803,44 @@ func TestUnwrap(t *testing.T) {
 	}
 
 }
+
+// -----------------------------------------------
+
+type MyStruct1 struct {
+	A string `json:"a"`
+}
+
+type MyStruct2 struct {
+	A string `json:"a"`
+	B string `json:"b,omitempty"`
+}
+
+func TestReadOmitEmptyJSON(t *testing.T) {
+	s := &MyStruct1{"hi"}
+	s2 := new(MyStruct2)
+
+	b := JSONBytes(s)
+	err := ReadJSONBytes(b, s2)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if s.A != s2.A {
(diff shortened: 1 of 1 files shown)


melekes (approved):

odeke-em (changes requested):
@ebuchman the test seems to be failing with
`reflect_test.go:842: Varint overflow`
which doesn't seem to have a dependence on if we use ",omitempty" or not, please see 
```diff
diff --git a/reflect_test.go b/reflect_test.go
index 965eb06..919292a 100644
--- a/reflect_test.go
+++ b/reflect_test.go
@@ -812,7 +812,7 @@ type MyStruct1 struct {
 
 type MyStruct2 struct {
 	A string `json:"a"`
-	B string `json:"b,omitempty"`
+	B string `json:"b"`
 }
 
 func TestReadOmitEmptyJSON(t *testing.T) {
@@ -835,6 +835,8 @@ func TestReadOmitEmptyBinary(t *testing.T) {
 	s2 := new(MyStruct2)
 
 	b := BinaryBytes(s)
+	bs := BinaryBytes(&MyStruct2{A: "hi"})
+	t.Logf("\nb: %x\nbs:%x\n", b, bs)
 	err := ReadBinaryBytes(b, s2)
 	if err != nil {
 		t.Fatal(err)
```
which fails the same, with
```shell
--- FAIL: TestReadOmitEmptyBinary (0.00s)
	reflect_test.go:839: 
		b: 0101026869
		bs:010102686900
	reflect_test.go:842: Varint overflow
FAIL
exit status 1
FAIL	github.com/tendermint/go-wire	0.010s
```

ebuchman:
Yep so this is expected behaviour of go-wire - it's not extensible like JSON or protobuf are. We should document that better.

PR #43, opened 2017-10-23
fix default encoding of time and bytes

ebuchman:
Prior to this, empty byte arrays were decoded as []byte{} rather than []byte(nil), and empty time.Time structs were decoded as some date in 1754, since time.UnixNano is undefined for values that cant be represented by int64, including the default time (https://golang.org/pkg/time/#Time.UnixNano)

### DIFF
--- a/time.go
+++ b/time.go
@@ -12,16 +12,32 @@ Writes nanoseconds since epoch but with millisecond precision.
 This is to ease compatibility with Javascript etc.
 */
 
+// WriteTime writes the number of nanoseconds, with millisecond resolution,
+// since January 1, 1970 UTC, to the Writer as an Int64.
+// NOTE: panics if the given time is less than January 1, 1970 UTC
 func WriteTime(t time.Time, w io.Writer, n *int, err *error) {
 	nanosecs := t.UnixNano()
 	millisecs := nanosecs / 1000000
-	WriteInt64(millisecs*1000000, w, n, err)
+	if nanosecs < 0 {
+		cmn.PanicSanity("can't encode times below 1970")
+	} else {
+		WriteInt64(millisecs*1000000, w, n, err)
+	}
 }
 
+// ReadTime reads an Int64 from the Reader, interprets it as
+// the number of nanoseconds since January 1, 1970 UTC, and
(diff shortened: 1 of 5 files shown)


time.go
```diff
@@ -12,16 +12,32 @@ Writes nanoseconds since epoch but with millisecond precision.
...
 
+// WriteTime writes the number of nanoseconds, with millisecond resolution,
+// since January 1, 1970 UTC, to the Writer as an Int64.
+// NOTE: panics if the given time is less than January 1, 1970 UTC
 func WriteTime(t time.Time, w io.Writer, n *int, err *error) {
 	nanosecs := t.UnixNano()
 	millisecs := nanosecs / 1000000
-	WriteInt64(millisecs*1000000, w, n, err)
+	if nanosecs < 0 {
+		cmn.PanicSanity("can't encode times below 1970")
```
odeke-em, line 22: This looks like it is a backwards incompatible change since structs that are serialized with the zero time value will now panic. Perhaps let's discuss this more.

And now we come to a rough idea here, but perhaps we should just encode nanoseconds and not worry about panicking with sub-millisecond precisionand then in ReadTime
```go
t := ReadInt64(r, n, err)
if t < 0 {
   return time.Time{}
}
return time.Unix(0, t)
```
What do you think?

I also have a naive question now: why do we have truncation of time to the millisecond when serializing but nowhere with time usage (at least in https://github.com/tendermint/tendermint), do we see:
```go
t.Round(time.Millisecond)
```
?

ebuchman: It's for compatibility with javascript, which can only handle MS.

This is indeed a breaking change. But the functionality is currently broken, ie. 

```
buf
WriteTime(time.Time{}, buf ...)
ReadTime(buf...) ! = time.Time{}
```

odeke-em: Oh, could we perhaps document the compatibility need with Javascript?

odeke-em: /cc @zramsay

ebuchman: see 3180c86

odeke-em: Awesome, thank you! That makes it a lot clearer.

PR #50, opened 2017-11-16
fix bug in uint32 / uint16 cut paste error

petabytestorage:


odeke-em (approved):
LGTM, thank you for all the work and for the quick update after the conversation offline.

PR #52, opened 2017-11-30
data: add Marshal and Unmarshal methods to Bytes

ebuchman:


melekes (approved):

data/bytes.go
```diff
@@ -30,6 +30,17 @@ var (
...
+// Marshal needed for protobuf compatibility
+func (b Bytes) Marshal() ([]byte, error) {
+	return b, nil
+}
+
+// Unmarshal needed for protobuf compatibility
+func (b *Bytes) Unmarshal(data []byte) error {
+	*b = data
+	return nil
+}
```
odeke-em, line 42: In the future let's add a compile time assertion to ensure that protobuf compatibility
so
```go
var _ protobuf.Marshaler = (*Bytes)(nil)
var _ protobuf.Unmarshaler = (*Bytes)(nil)
```
this way if some code changes after being edited by someone, they'll fix it instead of blocking others :)

PR #53, opened 2017-12-01
fix length bug - doubly incremented n

petabytestorage:
incremented in WriteTo and before. two places, same pointer address so same int incremented doubly incorrectly.

### DIFF
--- a/int_test.go
+++ b/int_test.go
@@ -8,6 +8,15 @@ import (
 
 const arch64 = uint64(^uint(0)) == ^uint64(0)
 
+func TestInt16(t *testing.T) {
+	buf := new(bytes.Buffer)
+	n, err := new(int), new(error)
+	WriteUint16(11, buf, n, err)
+	if *n != 2 {
+		t.Fatalf("Encoded uint16 expected 2 bytes got %d", *n)
+	}
+}
+
 // Returns true of overflow or underflow
 func checkIntSpill(i int64) bool {
 	return i == (int64)((int)(i))
--- a/int.go
+++ b/int.go
@@ -88,7 +88,6 @@ func ReadUint8(r io.Reader, n *int, err *error) uint8 {
 func WriteInt16(i int16, w io.Writer, n *int, err *error) {
 	var buf [2]byte
 	binary.BigEndian.PutUint16(buf[:], uint16(i))
-	*n += 2
 	WriteTo(buf[:], w, n, err)
 }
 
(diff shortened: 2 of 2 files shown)


melekes (approved):

int_test.go
```diff
@@ -8,6 +8,15 @@ import (
 
 const arch64 = uint64(^uint(0)) == ^uint64(0)
 
+func TestInt16(t *testing.T) {
```
odeke-em, line 11: @petabytestorage could we enumerate the same test for the respective *int* Writers and the expected values? That'll help us catch and prevent any regressions.

odeke-em (changes requested):
Just a request for more test enumerations for the other functions and bug fixes.

int_test.go
petabytestorage: i'm rewriting all this code in a better interface that eliminates this so i will leave it up to @ebuchman and @jaekwon how much effort that want to put into fixing this top-level int.go (depends how much it's used i guess in other places?) versus fixing the new one in nowriter/tmlegacy/... and indeed we are moving away from that interface and style so all these bugs will go away as we move into the `[]byte` format that @jaekwon has supported for some time. so i think fastest is just finish up the new good interfaces, add simple wrapper to support buf accumulation (i will be adding that soon) then we can move everything over to new safer interface and have all these bugs gone because we no longer manipulate int n byte count directly because we no longer use old legacy interface.  therefore my suggestion is merge my other pr and lets get finishing. https://github.com/tendermint/go-wire/pull/51

odeke-em: @petabytestorage sure, but the problem though is that we had that code in there and no one noticed those bugs, I believe enumerating those tests is a copy and paste and changing the sizes. If you'd like I could send a follow up PR to implement them. I agree, finish up as much.

petabytestorage: i accept that offer!  :ok_woman: ty @odeke-em

odeke-em: Cheers :), please ping me once this PR has been merged and I'll send them along.

petabytestorage: i think if you make a PR to be merged into this PR then i merge it it would clear your old blocking review and then you can approve. so save a step or two not sure tho about this github workflow gui thing.

ebuchman (approved):
OMG lol. Thanks Rudi.

int_test.go
petabytestorage: it's merge @odeke-em

PR #62, opened 2017-12-07
Add length-prefixed and unlength-prefixed array encoders to complete …

petabytestorage:
…the set  closes https://github.com/tendermint/go-wire/issues/59 and https://github.com/tendermint/go-wire/issues/58

### DIFF
--- a/nowriter/tmarrayencoder/tm_array_encoder_unlength_pure.go
+++ b/nowriter/tmarrayencoder/tm_array_encoder_unlength_pure.go
@@ -1,0 +1,119 @@
+package tmarrayencoder
+
+import (
+	"time"
+)
+
+type TMArrayEncoderUnlengthPure struct {
+}
+
+var _ TMArrayEncoderUnlength = TMArrayEncoderUnlengthPure{}
+
+//var basis tmenc
+func (a TMArrayEncoderUnlengthPure) EncodeBoolArray(b []bool) (ary []byte) {
+	for _, e := range b {
+		ary = append(ary, elementEncoder.EncodeBool(e)...)
+	}
+	return
+}
+
+func (a TMArrayEncoderUnlengthPure) EncodeFloat32Array(f []float32) (ary []byte) {
+	for _, e := range f {
+		ary = append(ary, elementEncoder.EncodeFloat32(e)...)
+	}
+	return
+}
+
+func (a TMArrayEncoderUnlengthPure) EncodeFloat64Array(f []float64) (ary []byte) {
+	for _, e := range f {
+		ary = append(ary, elementEncoder.EncodeFloat64(e)...)
+	}
+	return
+}
+
+func (a TMArrayEncoderUnlengthPure) EncodeInt8Array(i []int8) (ary []byte) {
+	for _, e := range i {
+		ary = append(ary, elementEncoder.EncodeInt8(e)...)
+	}
+	return
+}
+
+func (a TMArrayEncoderUnlengthPure) EncodeInt16Array(i []int16) (ary []byte) {
+	for _, e := range i {
+		ary = append(ary, elementEncoder.EncodeInt16(e)...)
+	}
+	return
+}
+
+func (a TMArrayEncoderUnlengthPure) EncodeInt32Array(i []int32) (ary []byte) {
+	for _, e := range i {
+		ary = append(ary, elementEncoder.EncodeInt32(e)...)
+	}
+	return
+}
+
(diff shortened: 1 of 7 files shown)


nowriter/tmarrayencoder/tm_array_encoder_unlength_pure.go
```diff
@@ -0,0 +1,119 @@
...
+import (
+	"time"
+)
+
+type TMArrayEncoderUnlengthPure struct {
+}
+
+var _ TMArrayEncoderUnlength = TMArrayEncoderUnlengthPure{}
+
+//var basis tmenc
```
melekes: looks like a leftover

melekes (approved):

nowriter/tmwire/tmwire_test.go
```diff
@@ -16,6 +17,22 @@ import (
...
+	encodedBytes := arrayRaw.EncodeUint16Array(simpleTest)
+	assert.Zero(t, bytes.Compare(encodedBytes, expected))
+}
+
+func TestArrayPrefix(t *testing.T) {
+	simpleTest := []uint16{0x1122, 0x3344}
+	expected := []byte{0x1, 0x2, 0x11, 0x22, 0x33, 0x44}
+	encodedBytes := arrayPrefixed.EncodeUint16Array(simpleTest)
+	assert.Zero(t, bytes.Compare(encodedBytes, expected))
+}
```
spacejam, line 35:
```
import "testing/quick"
...
func TestArray(t *testing.T) {                                                                                                           
  propertyTest := func(input []uint8) bool {                                                                                             
    truncatedInputLen := len(input) / 2 * 2                                                                                              
    truncatedInput := input[:truncatedInputLen]                                                                                          
                                                                                                                                         
    pairs := []uint16{}                                                                                                                  
    for i := 0; i+1 < truncatedInputLen; i += 2 {                                                                                        
      pair := uint16(truncatedInput[i]<<8) + uint16(truncatedInput[i+1])                                                                 
      pairs = append(pairs, pair)                                                                                                        
    }                                                                                                                                    
                                                                                                                                         
    encodedBytes := arrayRaw.EncodeUint16Array(pairs)                                                                                    
                                                                                                                                         
    return bytes.Compare(encodedBytes, truncatedInput) == 0                                                                                       
  }                                                                                                                                      
                                                                                                                                         
  assert.Nil(t, quick.Check(propertyTest, nil))                                                                                          
}
```

spacejam: caught a bug in my code, make sure the `bytes.Compare` compares against `truncatedInput` instead of the possibly longer `input`

spacejam: https://github.com/tendermint/go-wire/issues/63

ebuchman (approved):
Dope. What's the deal with the `PrefixStatus` method ?

petabytestorage:
@ebuchman it ensures that a TMArrayEncoder cannot implement both the Length and Unlength specialization interfaces simultaneously; it must indicate which one it implements. The type indicates the property of producing a self-delimiting prefix encoding. Since go doesn't allow method overloads I think it achieves our compile-time mutex purpose.

ebuchman:
Ah ok. 

Typically we use a method like `_assertTypeNameHere()`

eg. https://github.com/tendermint/go-wire/blob/master/data/base58/encoder.go#L21

This is fine too, but maybe we should converge on a common style for doing this

petabytestorage:
how does _assertTypeNameHere achieve the purpose of preventing people from simultaneously satisfying the requirements of two different interfaces that are supposed to be "either-or" ?  from my understanding that convention could introduce two new symbol names for two new methods each with different signatures... so it doesn't prevent the problem or am i missing something? @ebuchman

ebuchman:
you're 100% right! Thanks for clarifying that :)

ebuchman:
So maybe the convention should be to have a method like `_assert(InterfaceName)`. 

`PrefixStatus` just maybe isn't as clear as it could be about what's going on.

Anyways, just shaving yaks here ;)

petabytestorage:
Thanks for confirming and practicing the precise understanding. I agree that's appears to be a good candidate name and fine to me.  I'm happy to let you and the team arbitrate any such name better than the one I provided in another issue and maybe PR then i can standardize on it during the next mutually-exclusive (psuedo linear) type. For me it's just expressing the mental model (it's a bit unnatural since types are usually constructionist i.e. non-contradicting) as a simple logical formula at compile time that i appreciate and whatever name is most familiar to the team is best. @ebuchman

PR #178, opened 2018-06-13
Don't encode zero values in time

liamsi:
Skip over default values (zeroes) in `EncodeTime` (as proto3 does).


melekes (approved):

liamsi:
Thanks @melekes !

PR #179, opened 2018-06-13
do not encode empty structs, unless `amino:"write_empty"` is set

liamsi:
- [x] encoding & basic tests 
- [x] decoding with writeEmpty set to true
- [x] more tests, testing, fuzzing


jaekwon (approved):

liamsi:
Thanks for the review @jaekwon!

PR #180, opened 2018-06-14
only return error if prefix bytes are less than 4 bytes

melekes:
Fixes https://github.com/tendermint/tendermint/pull/1721#issuecomment-397221418


liamsi (approved):
LGTM

PR #190, opened 2018-06-18
Fix zero time decoding

### LINKED ISSUE #189: Fix failing tests on circleci before release v0.11.0
Decoding time sometimes fails because the 1st field might be skipped as with #178

I have a fix for this lined up, will push later.

https://circleci.com/gh/tendermint/go-amino/413?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

liamsi:
fix for changes in #178 (introduced a decoding bug for zero values)

resolves #189 

PTAL @xla @jaekwon

### DIFF
--- a/amino.go
+++ b/amino.go
@@ -15,8 +15,21 @@ import (
 // Global methods for global sealed codec.
 var gcdc *Codec
 
+// we use this time to init. a zero value (opposed to reflect.Zero which gives time.Time{} / 01-01-01 00:00:00)
+var zeroTime time.Time
+
+const (
+	unixEpochStr = "1970-01-01 00:00:00 +0000 UTC"
+	epochFmt     = "2006-01-02 15:04:05 +0000 UTC"
+)
+
 func init() {
 	gcdc = NewCodec().Seal()
+	var err error
+	zeroTime, err = time.Parse(epochFmt, unixEpochStr)
+	if err != nil {
+		panic("couldn't parse Zero value for time")
+	}
 }
 
 func MarshalBinary(o interface{}) ([]byte, error) {
@@ -87,9 +100,6 @@ const (
 	//Typ3_List       = Typ3(6)
 	//Typ3_Interface  = Typ3(7)
 )
-
-// we use this time to init. a zero value (opposed to reflect.Zero which gives time.Time{} / 01-01-01 00:00:00)
-var zeroTime, _ = time.Parse("2006-01-02 15:04:05 +0000 UTC", "1970-01-01 00:00:00 +0000 UTC")
 
 func (typ Typ3) String() string {
 	switch typ {
--- a/amino.go
+++ b/amino.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"reflect"
+	"time"
 )
 
 //----------------------------------------
@@ -86,6 +87,9 @@ const (
 	//Typ3_List       = Typ3(6)
 	//Typ3_Interface  = Typ3(7)
 )
+
+// we use this time to init. a zero value (opposed to reflect.Zero which gives time.Time{} / 01-01-01 00:00:00)
+var zeroTime, _ = time.Parse("2006-01-02 15:04:05 +0000 UTC", "1970-01-01 00:00:00 +0000 UTC")
 
 func (typ Typ3) String() string {
 	switch typ {
--- a/decoder.go
+++ b/decoder.go
@@ -65,7 +65,7 @@ func DecodeVarint(bz []byte) (i int64, n int, err error) {
 	i, n = binary.Varint(bz)
 	if n == 0 {
 		// buf too small
-		err = errors.New("varint buffer too small")
+		err = errors.New("buffer too small")
 	} else if n < 0 {
 		// value larger than 64 bits (overflow)
 		// and -n is the number of bytes read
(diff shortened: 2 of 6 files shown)


amino.go
```diff
@@ -87,6 +88,9 @@ const (
 	//Typ3_Interface  = Typ3(7)
 )
 
+// we use this time to init. a zero value (opposed to reflect.Zero which gives time.Time{} / 01-01-01 00:00:00)
+var zeroTime, _ = time.Parse("2006-01-02 15:04:05 +0000 UTC", "1970-01-01 00:00:00 +0000 UTC")
```
xla: It's not very likely but ignoring the error is not advisable. There are multiple ways to do this, one is to use a global and then have an init function which panics if the parse fails, or have a little wrapper like `mustParse` which panics in case of an error. The latter is desirable as use of `init` needs to be considered and initialisation order needs to be kept in mind.

Those strings would make for great constants.

decoder.go
```diff
@@ -198,54 +198,23 @@ func DecodeFloat64(bz []byte) (f float64, n int, err error) {
...
-		}
+	sec, n, err = decodeSeconds(&bz)
+	if err != nil {
+		return
 	}
-	// Actually read the Int32.
-	var nsec = int32(0)
-	nsec, _n, err = DecodeInt32(bz)
-	if slide(&bz, &n, _n) && err != nil {
+	nsec, err := decodeNanos(&bz, &n)
```
xla, line 213: The implicit return values make it hard to reason about what is returned at what point. Here for example is the returned `err` set correctly? If possible we should avoid them, good reasons to use implicit returns is for access in `defer`s.

decoder.go
```diff
@@ -258,6 +227,52 @@ func DecodeTime(bz []byte) (t time.Time, n int, err error) {
...
+		if slide(bz, &n, _n) && err != nil {
+			return 0, n, err
+		} else {
+			return sec, n, err
+		}
+	} else if fieldNum == 2 && typ == Typ3_4Byte {
+		// skip: do not slide, no error, will read again
+		return 0, n, nil
+	} else {
+		return 0, n, fmt.Errorf("expected field number 1 <8Bytes> or field number 2 <4Bytes> , got %v", fieldNum)
```
xla, line 251: Is there a general error strategy for amino? While totally out of scope of this PR, errors should be properly constructed (sentinels, or types) to give control to the caller. Esp. important with low-level libs.

decoder.go
```diff
@@ -258,6 +227,52 @@ func DecodeTime(bz []byte) (t time.Time, n int, err error) {
...
+	}
+}
+
+func decodeNanos(bz *[]byte, n *int) (int32, error) {
+	// Optionally decode field number 2 and Typ3 (4Byte).
+	//var fieldNum, typ, _n = uint32(0), Typ3(0x00), int(0)
```
xla: Maybe wec an rewrite this comment to show the layout in the byte slice, as an example check [this section](https://github.com/tendermint/go-amino#prefix-bytes-to-identify-the-concrete-type) from the README.

encoder.go
```diff
@@ -117,7 +117,6 @@ func EncodeFloat64(w io.Writer, f float64) (err error) {
 func EncodeTime(w io.Writer, t time.Time) (err error) {
 	var s = t.Unix()
 	var ns = int32(t.Nanosecond()) // this int64 -> int32 is safe.
-
```
xla: Any particular reason to remove the whitespace here?

time2_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDecodeTime(t *testing.T) {
+
```
xla: Unnecessary newline.

time2_test.go
```diff
@@ -0,0 +1,35 @@
...
+	type T struct {
```
xla: No need to have this uppercase, would be nice to give it a more descriptive title.

xla (approved):
:+1: :octocat: :shipit: :dango: 

LGTM, just some style and consistency comments.

decoder.go
liamsi: Yes, returns the correct err. But I agree with with the implicit returns. I did not want to change the public method here for the sake of consistency: all methods in this file have named return values/implicit returns.

decoder.go
liamsi: Agreed! But out of scope for this PR as you said (https://github.com/tendermint/go-amino/issues/191)

encoder.go
liamsi: nope... I think, I've added some debugging code which I deleted later (and with it that innocent line). Can reintroduce that.

decoder.go
liamsi: Thanks! I just forgot to delete this comment. I don't think there should be a general comment about how the byteslice will look here. It's very general (typ3AndFieldnum | actualBytesOfField) where typ3AndFieldnum is encoded via: (field_number << 3) | typ3 and not special. 
I'll delete the comment. A more public facing documentation should explain how time is encoded though.

amino.go
liamsi: Thanks! Very good suggestions 👍

time2_test.go
liamsi: adressed

time2_test.go
liamsi: adressed

liamsi:
Thanks for the timely and thorough review @xla!
If the fuzzer doesn't find any issue I'll merge today (cc @jaekwon)

decoder.go
xla: Understood, changing those won't affect anything on the caller side tho.

PR #196, opened 2018-06-20
Fix another time decoding issue

### LINKED ISSUE #189: Fix failing tests on circleci before release v0.11.0

liamsi:
This PR fixes the remaining cases (time in arrays, slices, pointers to time) where decoding time fails (if it was skipped over while encoding; introduced via #178).

Tests passed locally (incl. the fuzzer with no crashers for >1 hour). It still would be good to see circleci passing before this gets merged (see #200).

Thanks for the suggestions @jaekwon!

related to #189 #190 
resolves #189
Supersedes and closes #200

### DIFF
--- a/amino.go
+++ b/amino.go
@@ -444,3 +444,19 @@ func (cdc *Codec) MarshalJSONIndent(o interface{}, prefix, indent string) ([]byt
 	}
 	return out.Bytes(), nil
 }
+
+// Returns the default value of a type. For a time type or a pointer(s) to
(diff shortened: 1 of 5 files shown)


liamsi:
resolved via #201

PR #200, opened 2018-07-04
Fix develop: revert skipping over time.Time

liamsi:
This PR reverts skipping over time fields if they are empty #178 (and a related but non-complete fix #190). 

As you can see in both PRs circleci didn't run in these PRs (only after they got merged into develop). Hence, we didn't notice that this is going to break develop. See also below (circleci doesn't run again, tested only locally).
~~Can we change this behaviour please, so this won't happen again? (I don't have the necessary rights to change this) @ebuchman @jaekwon~~ Thanks Bucky!

There is also a WIP PR which will resolve these issues #196. Together with the 2 reverted commits it will have the effect that empty/default time-fields are "skipped" not encoded at all.


liamsi:
resolved via #201

PR #224, opened 2018-09-06
proto3 compatibility: use uvarint encoding in time instead of fixed sized

### LINKED ISSUE #208: Proposal: make time varint encoded (instead of fixed length)
In protobuf there is https://github.com/golang/protobuf/blob/14aad3d5ea4c323bcd7a2137e735da24a76e814c/ptypes/timestamp/timestamp.pb.go#L99-L108
where seconds and nanos are encoded as `varint`. See also the corresponding proto file / message: 
https://github.com/golang/protobuf/blob/14aad3d5ea4c323bcd7a2137e735da24a76e814c/ptypes/timestamp/timestamp.proto#L121-L133

In amino, we currently encode time as fixed length (signed) ints: 
https://github.com/tendermint/go-amino/blob/2106ca61d91029c931fd54968c2bb02dc96b1412/encoder.go#L123-L136

I think it could be worthwhile to change this behaviour. Then, people working on amino in other languages, or interefacing via vanilla protobuf could just use proto's encoding / types.

### LINKED ISSUE #223: Encoding for time should be the same as Google.Protobuf.WellKnownTypes.Timestamp
There is some pretense that we are the same as the Google.Protobuf.WellKnownTypes.Timestamp but turns out we're not.

See the note in https://github.com/tendermint/tendermint/blob/develop/types/proto3/block.proto#L43

And the code in Amino: https://github.com/tendermint/go-amino/blob/faa6e731944e2b7b6a46ad202902851e8ce85bee/encoder.go#L128-L139

We should fix amino so this aligns properly, unless there is a very good reason not to.

liamsi:
~~WIP: fuzzer found a crasher, still need to investigate if this is related!~~
The fuzzer just timed out, no real crashers.

resolves #208 
resolves #223


liamsi:
This is actually a bit confusing:
proto3's Timestamp message looks like this:

```proto
message Timestamp {
  int64 seconds = 1; // this is proto3's int64
  int32 nanos = 2;
}
```
which translate into this go code:
```go
type Timestamp struct {
	Seconds int64 `protobuf:"varint,1,opt,name=seconds,proto3" json:"seconds,omitempty"`
	Nanos   int32    `protobuf:"varint,2,opt,name=nanos,proto3" json:"nanos,omitempty"`
}
```
In protobuf `int32` and `int64` are casted to `uint64` and encoded as unsigned varint. In amino we encode golang's `int32` as varint (*not* unsigned). We treat it as proto's `sint32` which is encoded differently (see [scalar types]). This isn't a problem but probably needs to be documented somewhere. 

I've updated amino's time encoding to match with proto's behaviour (will push shortly).

[scalar types]: https://developers.google.com/protocol-buffers/docs/proto#scalar

ebuchman:
@melekes can you help review this?

encoder.go
```diff
@@ -111,9 +112,25 @@ func EncodeFloat64(w io.Writer, f float64) (err error) {
 	return EncodeUint64(w, math.Float64bits(f))
 }
 
+const (
+	// seconds of 01-01-0001
+	minSeconds = -62135596800
+	// seconds of 10000-01-01
+	maxSeconds = 253402300800
```
ValarDragon, line 119: Is this the standard upperbound? Just seems kinda weird since its not near a power of two. Doesn't block the PR though

decoder.go
```diff
@@ -261,15 +262,20 @@ func decodeNanos(bz *[]byte, n *int) (int32, error) {
 	if err != nil {
 		return 0, err
 	}
-	if fieldNum == 2 && typ == Typ3_4Byte {
+	if fieldNum == 2 && typ == Typ3_Varint {
 		slide(bz, n, _n)
 		_n = 0
 		// Actually read the Int32.
```
ValarDragon: Outdated comment

tests/proto3/proto3_compat_test.go
```diff
@@ -173,8 +183,154 @@ func TestProto3CompatPtrsRoundtrip(t *testing.T) {
...
+	// decoding; protobuf defaults to nil here (see the following lines below):
+	assert.NoError(t, err)
+	assert.Equal(t, goAminoGotTime{T: &epoch}, *res)
+	pbRes := p3.ProtoGotTime{}
+	err = proto.Unmarshal(pb, &pbRes)
+	assert.NoError(t, err)
+	assert.Equal(t, p3.ProtoGotTime{T: nil}, pbRes)
+}
+
+func TestProto3CompatTimestampNow(t *testing.T) {
```
ValarDragon, line 221: Can this be table driven?

ValarDragon (approved):
Looks good to me, just a few minor comments!

encoder.go
liamsi: This constant is straight from the code that converts between `time.Time` and the wellknown type `Timestamp`: https://github.com/golang/protobuf/blob/ddf22928ea3c56eb4292a0adbbf5001b1e8e7d0d/ptypes/timestamp.go#L48-L50

decoder.go
liamsi: Thanks, good catch! Removed.

tests/proto3/proto3_compat_test.go
liamsi: That would be nice. But it's basically one test-case that documents what works how.

liamsi:
Merged! Thanks for the review @ValarDragon

decoder.go
```diff
@@ -238,20 +232,27 @@ func decodeSeconds(bz *[]byte) (int64, int, error) {
...
 			return 0, n, err
-		} else {
-			return sec, n, err
 		}
-	} else if fieldNum == 2 && typ == Typ3_4Byte {
+		// if seconds where negative before casting them to uint64, we yield
+		// the original signed value:
+		res := int64(sec)
+		if res < minSeconds || res >= maxSeconds {
+			return 0, n, InvalidTimeErr(fmt.Sprintf("seconds have to be > %d and < %d, got: %d",
```
melekes, line 246: `seconds have to be >= %d`

encoder.go
```diff
@@ -123,22 +140,33 @@ func EncodeTime(w io.Writer, t time.Time) (err error) {
...
 		}
 	}
 	// skip if default/zero value:
 	if ns != 0 {
-		err = encodeFieldNumberAndTyp3(w, 2, Typ3_4Byte)
+		// do not encode if not in interval [0, 999999999]
+		if ns < 0 || ns > maxNanos {
+			// we could as well panic here:
+			// time.Time.Nanosecond() guarantees nanos to be in [0, 999,999,999]
+			return InvalidTimeErr(fmt.Sprintf("nanosecons have to be >= 0 and <= %v, got: %d",
```
melekes, line 162: `nanoseconds`

liamsi: Thanks. I'll address this and above comment in #233

PR #225, opened 2018-09-08
No longer allocate space for uvarint size

ValarDragon:
In an area within the mempool, we ended not going with amino's implementation to due to allocation on every call. Its an easy fix, and I've added test cases to prove correctness.


liamsi (approved):
excellent, small PR! Thanks @ValarDragon

PR #234, opened 2018-10-22
Release v0.13.0

liamsi:
Changelog: https://github.com/tendermint/go-amino/blob/develop/CHANGELOG.md#0130-october-15-2018
Something odd is going on with the commit history here (there are commits that are already in master, the diff looks correct though). I'll need to fix that first.


ebuchman (approved):
Very cursory review, but looks like all the individual PRs that make this up were reviewed.

liamsi:
Thanks @ebuchman! There is sth wrong with the commit history though. Probably merged an outdated master into develop (?) or did some weird squash-foo. I'll fix this later / tomorrow morning (the latest).

liamsi:
OK, I think what is going on: the commits we are seeing actually never made it into master (because we squashed the changes into master and hence the commits still show up here). Merging now.

Issue #237, opened 2018-10-24
Default encoding of unsigned ints


liamsi:
Summary of proposed changes: https://github.com/tendermint/tendermint/issues/2684

tl;dr: Current encoding in amino: amino's `int32` matches protobuf's `zigzag32` (which is an `sint32` in the proto file and which generates a `int32` but `zigzag32` encoded in go-lang)
Desired behaviour: default e.g., `int32` encoding to protobuf's `varint` encoding (which is a `int32` in the proto file and which also generates a `int32` in go-lang but `varint` encoded)

liamsi:
closed via #238

PR #238, opened 2018-10-25
Make default int encoding varint (not zigzag)

liamsi:
First stab on #237 
related to https://github.com/tendermint/tendermint/issues/2682

I propose to do a followup PR to add back the zigzag encoding option (via a tag).


binary-decode.go
```diff
@@ -115,11 +115,12 @@ func (cdc *Codec) decodeReflectBinary(bz []byte, info *TypeInfo, rv reflect.Valu
...
 			rv.SetInt(num)
 		} else {
-			num, _n, err = DecodeVarint(bz)
+			var unum uint64
+			unum, _n, err = DecodeUvarint(bz)
 			if slide(&bz, &n, _n) && err != nil {
 				return
 			}
-			rv.SetInt(num)
+			rv.SetInt(int64(unum))
```
melekes: should we also check that `unum <= math.MaxInt64` ?

melekes (approved):

binary-decode.go
liamsi: Thanks. Yes, you are right, we should!

binary-decode.go
```diff
@@ -115,11 +115,12 @@ func (cdc *Codec) decodeReflectBinary(bz []byte, info *TypeInfo, rv reflect.Valu
 			}
 			rv.SetInt(num)
 		} else {
-			num, _n, err = DecodeVarint(bz)
+			var unum uint64
```
ValarDragon: This should not be called a unum. Unums are a cool proposal to replace floats. https://en.wikipedia.org/wiki/Unum_(number_format).

binary-decode.go
liamsi: Oh actually, this wouldn't make any sense in this case. Any negative value num will certainly be > math.MaxInt64 (all negative nums get encoded "after" MaxInt64). But checks make sense in the cases of `int32` (and depending on the platform, also for `int`). I'll add those.

binary-decode.go
liamsi: Universal numbers ... Thanks! Learned something new. I'll rename to `u64` or sth like this.

liamsi:
Addressed all comments and updated the changelog for another release. Want to have another look? Or good to merge @melekes @ValarDragon

binary-decode.go
```diff
@@ -12,6 +13,16 @@ import (
 //----------------------------------------
 // cdc.decodeReflectBinary
 
+var (
+	ErrOverflowInt = errors.New("encoded integer value value overflows int(32)")
```
melekes: value value

Issue #239, opened 2018-10-26
Reintroduce zigzag encoding via tag


liamsi:
As per #238 `int`s, `int32`s, and `int64`s aren't zigzag encoded by default. Add a tag such that developers can still zigzag encode. If they need to encode ints which often contain negative numbers this is more efficient.

TODOs:
 - [ ] add a tag
 - [ ] add back tests (see: https://github.com/tendermint/go-amino/pull/240#discussion_r228562934)

Zigzag encoding is used in proto3 for encoding `sint(32|64)`.
References: 
 - https://developers.google.com/protocol-buffers/docs/encoding?csw=1#types
 - https://developers.google.com/protocol-buffers/docs/proto3#scalar

liamsi:
removed from the [proto milestone](https://github.com/tendermint/go-amino/milestone/1): My understanding is, we require everything amino encodes can be represented as a protobuf message and not that amino supports every proto feature.

PR #240, opened 2018-10-26
Release 0.14.0

liamsi:


tests/proto3/proto3_compat_test.go
```diff
@@ -52,7 +56,8 @@ func TestFixed32Roundtrip(t *testing.T) {
 }
 
 func TestVarintZigzagRoundtrip(t *testing.T) {
-	// amino varint (int) <-> protobuf zigzag32 (int32)
+	t.Skip("zigzag encoding isn't default anymore for (unsigned) ints")
```
ebuchman, line 59: TODO add tags and test this ?

ebuchman (approved):

tests/proto3/proto3_compat_test.go
liamsi: Added an issue for this: #239 
~~I'll add a todo and a link to the issue (in a follow up PR), thanks!~~
I'll link you comment in that issue so we won't forget about it instead. Still Thanks :-)

PR #242, opened 2018-11-05
Make amino build on 32bit architectures

liamsi:
Some constants in encoder.go are interpreted as `int` by default and they overflow ints on 32bit machines: [encoder.go](https://github.com/tendermint/go-amino/blob/6dcc6ddc143e116455c94b25c1004c99e0d0ca12/encoder.go#L118-L122)
Similarly some values that seed the fuzzer: [here](https://github.com/tendermint/go-amino/blob/6dcc6ddc143e116455c94b25c1004c99e0d0ca12/tests/fuzz/binary/init-corpus/main.go#L51) for instance.

The first makes tendermint builds fail on certain architectures. This PR is a quick-fix for this issue. 
The changes made in this PR were tested (and fuzzed) with a i686 docker image (and locally for 64bit).

### DIFF
--- a/encoder.go
+++ b/encoder.go
@@ -119,7 +119,7 @@ const (
 	// seconds of 01-01-0001
 	minSeconds int64 = -62135596800
 	// seconds of 10000-01-01
-	maxSeconds = int64(253402300800)
+	maxSeconds int64 = 253402300800
 
 	// nanos have to be in interval: [0, 999999999]
 	maxNanos = 999999999
--- a/encoder.go
+++ b/encoder.go
@@ -117,7 +117,7 @@ func EncodeFloat64(w io.Writer, f float64) (err error) {
 
 const (
 	// seconds of 01-01-0001
-	minSeconds = int64(-62135596800)
+	minSeconds int64 = -62135596800
 	// seconds of 10000-01-01
 	maxSeconds = int64(253402300800)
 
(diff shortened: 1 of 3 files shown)


encoder.go
```diff
@@ -117,9 +117,9 @@ func EncodeFloat64(w io.Writer, f float64) (err error) {
 
 const (
 	// seconds of 01-01-0001
-	minSeconds = -62135596800
+	minSeconds = int64(-62135596800)
```
melekes: you can also do

```suggestion
	minSeconds int64 = -62135596800
```

, which I prefer slightly better

encoder.go
```diff
@@ -117,9 +117,9 @@ func EncodeFloat64(w io.Writer, f float64) (err error) {
...
 	// seconds of 10000-01-01
-	maxSeconds = 253402300800
+	maxSeconds = int64(253402300800)
```
melekes:
```suggestion
	maxSeconds int64 = 253402300800
```

melekes (approved):
🍇

greg-szabo (approved):
Looks good to me, but weren't there similar settings in `decoder.go`? Seems that only `encoder.go` was fixed.

liamsi:
Thanks for the review @greg-szabo. I think we were seeing errors in encoder.go because it was using that same variables defined in decoder.go. After #243 gets approved, I'll tag another release, which should make the arm/32bit builds work again.

PR #246, opened 2018-11-12
 fixed issue for 32 bit compile

Lunagram:


encoder.go
```diff
@@ -143,7 +143,7 @@ func EncodeTime(w io.Writer, t time.Time) (err error) {
 	if s != 0 {
 		if s < minSeconds || s >= maxSeconds {
 			return InvalidTimeErr(fmt.Sprintf("seconds have to be >= %d and < %d, got: %d",
-				minSeconds, maxSeconds, s))
+				int64(minSeconds), int64(maxSeconds), s))
```
liamsi, line 146: These constants are already of type `int64` (as per #242): https://github.com/tendermint/go-amino/blob/b2c3df8b60381e5fe0d6255171a0de7af3d2d65c/encoder.go#L119-L122

These casts should be redundant.

PR #259, opened 2019-03-20
Update link in README.md

harshad1ledger:
The section in tendermint docs regarding amino encoding seems to be moved to a new location. Updating it...


liamsi:
Thansk a lot @harshad1ledger for contributing to amino. This was fixed in https://github.com/tendermint/go-amino/pull/252  (on develop) and will be contained in the next release.

Issue #204, opened 2018-07-17
Wrap basic go-lang types (e.g bytes) in a message while de/encoding 


liamsi:
**(Final?) Update:** Instead of disallowing direct encoding of basic types, the SDK team demands the same convenience of amino as before. Meaning they still want to be able to encode basic types without wrapping them in a message / struct. 

The only alternative to still fully achieve protobuf compatibility I can think of: Amino internally when encoding does the wrapping for you. E.g. it will map `type SomeDefinedType []byte` to 
```proto
message BytesValue {
  // The bytes value.
  bytes value = 1;
}
```
(https://github.com/protocolbuffers/protobuf/blob/b4f193788c9f0f05d7e0879ea96cd738630e5d51/src/google/protobuf/wrappers.proto#L120-L123)

Same when decoding: it reads into such helper struct and gets the value from the field.

-----

**Update**/ tl;dr: Amino should not allow direct serialization of bytes nor of any built-in basic types (e.g. []bytes that aren't a field in a parenting struct). In protobuf they would need to be wrapped in a `message SomeMsg{ bytes actual_bytes = 1; }`, another example is: https://github.com/protocolbuffers/protobuf/blob/78ca77ac8799f67fda7b9a01cc691cd9fe526f25/src/google/protobuf/wrappers.proto#L117-L123 
Tendermint as well as the SDK use this feature of amino and avoid wrapping `[]bytes` in a `struct` and this could be changed independent from the amino API. An example is given below. Note that there are probably more places like this though (not only in secret connection).

The only downside is that this will add some overhead on the wire. The wrapping message adds (at least) two bytes for instance.

------
Original issue description: Amino API could be more restrictive

While working on the (de)serialization in Rust in secret connection / for kms I've noticed two ways how we use amino which might make it a bit tricky to implement amino deserialization in other languages. I wanted to keep track of them here (either to properly document this later or to discuss to make this easier):

1) we allow to enc / dec. raw bytes; sth. that protobuf doesn't allow to do (everything is always encapsulated in a message and has a field number); see for example how we use amino in secret connection: https://github.com/tendermint/tendermint/blob/2aa2b63cadc42cca1071c36adfd2f2ce14e1aa8f/p2p/conn/secret_connection.go#L199-L213
We might consider to not allow to use amino like this (on the other hand it might also actually be  useful to save that extra byte that would be necessary to enc. the fieldnum and type?).

[...]

PR #270, opened 2019-05-15
Wrap basic types in a struct

liamsi:
ref #204

Note: Instead of switching over all possible cases via reflection (and maybe recursively), this simply adds the bytes. This PR just emulates a wrapping struct.


amino.go
```diff
@@ -203,21 +211,43 @@ func (cdc *Codec) MarshalBinaryBare(o interface{}) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	err = cdc.encodeReflectBinary(buf, info, rv, FieldOptions{BinFieldNum: 1}, true)
-	if err != nil {
-		return nil, err
+	// in the case of of a repeated struct (e.g. type Alias []SomeStruct),
+	// we do not need to prepend with `(field_number << 3) | wire_type` as this
+	// would need to be done for each struct and not only for
```
liamsi:
```suggestion
	// would need to be done for each struct and not only for the first.
```

amino.go
```diff
@@ -387,10 +444,6 @@ func (cdc *Codec) MarshalJSON(o interface{}) ([]byte, error) {
 
 	// disfix wrapper continued...
 	if info.Registered {
-		// Part 2:
-		if err != nil {
-			return nil, err
-		}
```
liamsi, line 393: Not sure what this was supposed to do ... The err would always have been nil here.

json_test.go
```diff
@@ -172,24 +172,8 @@ type innerFP struct {
 }
 
 func TestUnmarshalMap(t *testing.T) {
-	binBytes := []byte(`dontcare`)
 	obj := new(map[string]int)
 	cdc := amino.NewCodec()
-	// Binary doesn't support decoding to a map...
-	// TODO: move out binary tests from json_test.go ...
-	assert.Panics(t, func() {
```
liamsi, line 180: These were not deleted but moved into binary_test.go

amino.go
```diff
@@ -2,21 +2,29 @@ package amino
...
+
+	// we use this time to init. a zero value (opposed to reflect.Zero which gives time.Time{} / 01-01-01 00:00:00)
+	zeroTime time.Time
 
-// we use this time to init. a zero value (opposed to reflect.Zero which gives time.Time{} / 01-01-01 00:00:00)
-var zeroTime time.Time
+	// NotPointerErr is thrown when you call a method that expects a pointer, e.g. Unmarshal
+	NotPointerErr = errors.New("expected a pointer")
+
+	UnknownBasicTypeErr = fmt.Errorf("got unknown type: ")
```
jordaaash: Is this error used anywhere?

Screen Shot 2019-05-22 at 12 41 02 PM

amino.go
```diff
@@ -2,21 +2,29 @@ package amino
...
+	// Global methods for global sealed codec.
+	gcdc *Codec
+
+	// we use this time to init. a zero value (opposed to reflect.Zero which gives time.Time{} / 01-01-01 00:00:00)
+	zeroTime time.Time
 
-// we use this time to init. a zero value (opposed to reflect.Zero which gives time.Time{} / 01-01-01 00:00:00)
-var zeroTime time.Time
+	// NotPointerErr is thrown when you call a method that expects a pointer, e.g. Unmarshal
+	NotPointerErr = errors.New("expected a pointer")
```
jordaaash, line 24: I'm not familiar with stdlib's `errors` vs. `pkg/errors` vs. `fmt.Errorf`, but I saw this dependency isn't used elsewhere. Worth adding a new library for?

amino.go
liamsi: Good catch! That's a relict from another approach (actually wrapping each basic type in a struct instead of just adding the necessary bytes). Removed.

amino.go
liamsi: pkg/errors allows you to wrap the error with some context instead of just passing it on. This helps callers to understand exactly where the error occurred. Although you can achieve sth. similar with fmt.Errorf, I think error.Wrap is slightly more convenient (and it also adds the call stack if you want to).

amino.go
Liamsi: This is an excellent read on errors and touches the different error packages too: https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully

amino.go
liamsi: This is an excellent read on errors and touches the different error packages too: https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully

amino.go
```diff
@@ -340,6 +371,28 @@ func (cdc *Codec) UnmarshalBinaryBare(bz []byte, ptr interface{}) error {
...
+	isRepeatedStruct := info.Type.Kind() == reflect.Slice && info.Type.Elem().Kind() == reflect.Struct
+	isPointerToStruct := isPtr && rv.Elem().Kind() != reflect.Struct
+	if rv.Kind() != reflect.Struct && !isRepeatedStruct && len(bz) > 0 && (rv.Kind() != reflect.Interface) && !isPointerToStruct {
+
+		fnum, typ, n, err := decodeFieldNumberAndTyp3(bz)
+		if err != nil {
+			return errors.Wrap(err, "could not decode field number and type")
+		}
+		if fnum != 1 {
+			fmt.Println("What is it?", rv.Kind())
```
liamsi: Forgot to delete this line.

binary-encode.go
```diff
@@ -413,27 +413,12 @@ func (cdc *Codec) encodeReflectBinaryStruct(w io.Writer, info *TypeInfo, rv refl
...
-					return
-				}
-				lAfterValue := buf.Len()
-
-				if !frvIsPtr && !fopts.WriteEmpty && lBeforeValue == lAfterValue-1 && buf.Bytes()[buf.Len()-1] == 0x00 {
-					// rollback typ3/fieldnum and last byte if
-					// not a pointer and empty:
-					buf.Truncate(lBeforeKey)
-				}
-
```
liamsi, line 436: This logic was moved to `writeFieldIfNotEmpty` as it is now used in two places (when wrapping a basic type as a field of a struct).

alexanderbez (approved):
ACK on first passthrough.

amino.go
```diff
@@ -340,6 +371,26 @@ func (cdc *Codec) UnmarshalBinaryBare(bz []byte, ptr interface{}) error {
 		}
 		bz = bz[4:]
 	}
+	isRepeatedStruct := info.Type.Kind() == reflect.Slice && info.Type.Elem().Kind() == reflect.Struct
```
alexanderbez: Maybe we ought to have an auxiliary function given an `info` that returns a bool if it's a repeated struct.

liamsi: Changed. Also, added another test case and a few related changes in 8764911

liamsi:
I just found another edge-case where this is not complete: when you have an interface and the concrete implementation is a type alias. Will add a test-case and fix shortly.

alexanderbez:
What's the best way to fully test this for all corner cases @Liamsi ?

liamsi:
More extensive & more coverage via unit-tests (and maybe extend the fuzzing code too). This would also have been caught, if one for instance ran tendermint / gaia using the modified version in this PR (b/c pub-keys are interfaces and their concrete implementations are type aliases).

I think, we can merge this PR as is. I'll do the change mentioned above in a follow-up PR ([issue](https://github.com/tendermint/go-amino/issues/273)). Before, we do a release, it would be great if we ran and tested it with tendermint and gaia before (additionally to the fuzzer and more unit tests).

alexanderbez:
Not a problem, I can easily test that out @Liamsi

Issue #273, opened 2019-05-28
Follow-up to #270: support registered type aliases


liamsi:
Write test and fix: https://github.com/tendermint/go-amino/pull/270#issuecomment-496292609

PR #274, opened 2019-05-28
Follow up PR #270: registered type aliases

liamsi:
ref #273


binary-decode.go
```diff
@@ -358,6 +359,27 @@ func (cdc *Codec) decodeReflectBinaryInterface(bz []byte, iinfo *TypeInfo, rv re
...
+		}
+		if fnum != 1 {
+			return n, fmt.Errorf("expected field number: 1; got: %v", fnum)
+		}
+		typWanted := typeToTyp3(cinfo.Type, FieldOptions{})
+		if typ != typWanted {
+			return n, fmt.Errorf("expected field type %v for # %v of %v, got %v",
+				typWanted, fnum, cinfo.Type, typ)
+		}
+		slide(&bz, &n, nFnumTyp3)
```
liamsi, line 381: This is the main change in this PR. Very similar to the changes done in #270 but for interfaces.

alexanderbez (approved):
ACK 👍

PR #282, opened 2019-08-14
faster VarintSize()

rickyyangz:
As I tested on my laptop, 

the original one costs:
```
Benchmark_VarintSize-12    	300000000	         6.21 ns/op
```

the optimized one costs:
```
Benchmark_VarintSize2-12    	2000000000	         0.52 ns/op
```

### DIFF
--- a/encoder.go
+++ b/encoder.go
@@ -42,9 +42,7 @@ func EncodeVarint(w io.Writer, i int64) (err error) {
 }
 
 func VarintSize(i int64) int {
-	var buf [10]byte
-	n := binary.PutVarint(buf[:], i)
-	return n
+	return UvarintSize(uint64((uint64(i) << 1) ^ uint64(i>>63)))
 }
 
 //----------------------------------------


encoder.go
```diff
@@ -42,9 +42,7 @@ func EncodeVarint(w io.Writer, i int64) (err error) {
 }
 
 func VarintSize(i int64) int {
-	var buf [10]byte
-	n := binary.PutVarint(buf[:], i)
-	return n
+	return UvarintSize(uint64((uint64(i) << 1) ^ uint64(i >> 63)))
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	return UvarintSize(uint64((uint64(i) << 1) ^ uint64(i>>63)))
```

melekes (approved):

alexanderbez (approved):
ACK -- do we need to handle any cases when i < 0?

ebuchman:
Awesome, but 

1) can we get a test for this, please?
2) is this even used anywhere?

PR #291, opened 2019-10-03
Getters for data on Codec

jackzampolin:
Expose read on typeInfos data to allow for client side code generation. Allowing users read access to these types would allow for generation of (for example) typescript definitions.

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -147,6 +147,18 @@ func NewCodec() *Codec {
 	return cdc
 }
 
+func (cdc *Codec) TypeInfos() map[reflect.Type]*TypeInfo {
+	return cdc.typeInfos
+}
+
+func (cdc *Codec) TypeInfosByName() map[string]*TypeInfo {
+	return cdc.nameToTypeInfo
+}
+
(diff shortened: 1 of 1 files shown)


aaronc (commented):
Looks like there might be a bit more involved in exposing these as I see a mutex in a bunch of the other methods. See: https://github.com/tendermint/go-amino/blob/master/codec.go#L259. Also I think we want to expose interfaces as well.

jackzampolin:
I believe that the mutexs don't matter if the codec is `locked`. Also can you point out where we can expose those interfaces?

ethanfrey:
What about accepting an interface in `RegisterCodec`, instead of a concrete struct (all throughout the sdk code).

Then one could pass in a `RecordingCodec` implementation that serializes all calls to some file, and if you pass that to the root `RegisterCodec` function in your app, you will get an output of all concrete types and interfaces.

hschoenburg:
just following up here. This would be a huge help for a cosmos typescript client lib. Not sure I know how to implement properly but with some guidance im happy to try.

PR #292, opened 2019-10-03
Expose types

jackzampolin:
Replaces https://github.com/tendermint/go-amino/pull/291

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -145,6 +145,18 @@ func NewCodec() *Codec {
 	return cdc
 }
 
+func (cdc *Codec) TypeInfos() map[reflect.Type]*TypeInfo {
+	return cdc.typeInfos
+}
+
+func (cdc *Codec) TypeInfosByName() map[string]*TypeInfo {
+	return cdc.nameToTypeInfo
+}
+
+func (cdc *Codec) DisFixToTypeInfo() map[DisfixBytes]*TypeInfo {
+	return cdc.disfixToTypeInfo
+}
+
 // This function should be used to register all interfaces that will be
 // encoded/decoded by go-amino.
 // Usage:


codec.go
```diff
@@ -145,6 +145,18 @@ func NewCodec() *Codec {
 	return cdc
 }
 
+func (cdc *Codec) TypeInfos() map[reflect.Type]*TypeInfo {
```
alexanderbez, line 148: I don't have too much context here, but simply returning maps on a reference type is somewhat dangerous as the caller can unexpectedly modify the reference (i.e. the codec). Instead, return a copy of the map.

alexanderbez (changes requested):

codec.go
alexanderbez: Also, let's add simple godocs.

ebuchman:
Can we open an issue for this please?

Also note there is the `Codec.PrintTypes` method which outputs a pretty table like: https://github.com/tendermint/tendermint/blob/master/docs/spec/blockchain/encoding.md#public-key-cryptography. Ie that table is the output of https://github.com/tendermint/tendermint/blob/master/crypto/encoding/amino/encode_test.go#L56

I'm not sure exactly the use case here, but if you don't need it programmatically, hopefully printing the table is sufficient. Otherwise you will need to grab the mutex and copy the map, as bez pointed out

jackzampolin:
This was some quick hacking I did one afternoon to try to expose the types tracked by amino in a `json` format so that we could turn them easily into typescript definitions. The `codec.PrintTypes` function doesn't give enough info for this. This PR should be in `draft` and thats my bad.

I believe that Hans is working on this so I would like to check with him before closing/doing more work.

ebuchman:
Oh cool - would a method that just dumps JSON be sufficient then? Might be easier than trying to programmatically export the mutex protected maps ...

jackzampolin:
@ebuchman I think it would. How to do that with the existing code? This also could be useful for generating `.proto` types.

PR #296, opened 2019-11-11
Add more linting

tac0turtle:
- added more linting

Signed-off-by: Marko Baricevic <marbar3778@yahoo.com>

### DIFF
--- a/amino.go
+++ b/amino.go
@@ -169,12 +169,15 @@ func (cdc *Codec) MarshalBinaryLengthPrefixed(o interface{}) ([]byte, error) {
 // MarshalBinaryLengthPrefixedWriter writes the bytes as would be returned from
 // MarshalBinaryLengthPrefixed to the writer w.
 func (cdc *Codec) MarshalBinaryLengthPrefixedWriter(w io.Writer, o interface{}) (n int64, err error) {
-	// var bz, _n = []byte(nil), int(0)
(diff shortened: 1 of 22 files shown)


binary-decode.go
```diff
@@ -30,7 +30,9 @@ const (
 // CONTRACT: rv.CanAddr() is true.
 func (cdc *Codec) decodeReflectBinary(bz []byte, info *TypeInfo,
 	rv reflect.Value, fopts FieldOptions, bare bool) (n int, err error) {
+
 	if !rv.CanAddr() {
+		fmt.Println(bz, "bz")
```
melekes: ???

melekes (approved):
👍

tessr (approved):

amino.go
```diff
@@ -331,7 +331,7 @@ func (cdc *Codec) UnmarshalBinaryLengthPrefixedReader(r io.Reader, ptr interface
 	}
 	l = int64(u64)
 	if l < 0 {
-		_ = errors.Errorf(
+		_ = errors.Errorf( //nolint:errcheck
```
melekes, line 334: what is this?)

tac0turtle: I was looking at this as well, with amino being a bit touchy I didn't want to change to much stuff in this PR so I left it

melekes: Right. But this code does not do anything, so maybe we should comment it completely.

PR #306, opened 2020-02-16
Fix bug in function decodeReflectJSONStruct

NicolasMahe:
This PR fixes a bug in the function `decodeReflectJSONStruct`.

To decode each field of the struct, this function is calling `decodeReflectJSON`. The issue was instead of using each `field.FieldOptions`, it was passing its parameter `fopts FieldOptions`:
https://github.com/tendermint/go-amino/blob/59d50ef176f686dc2c0c150e85a9381a78b9978a/json-decode.go#L416

The binary version `decodeReflectBinaryStruct` is already using each `field.FieldOptions`, so I just apply the same logic to the json version.
https://github.com/tendermint/go-amino/blob/59d50ef176f686dc2c0c150e85a9381a78b9978a/binary-decode.go#L867

I also add a new `Unsafe` struct to `json_test.go` to unit test the issue. Without the fix, the tests are failing.

---

Here is the link to the failed CI test before the fix: https://app.circleci.com/jobs/github/tendermint/go-amino/670

### DIFF
--- a/json-decode.go
+++ b/json-decode.go
@@ -362,7 +362,7 @@ func (cdc *Codec) decodeReflectJSONSlice(bz []byte, info *TypeInfo, rv reflect.V
 }
 
 // CONTRACT: rv.CanAddr() is true.
-func (cdc *Codec) decodeReflectJSONStruct(bz []byte, info *TypeInfo, rv reflect.Value, fopts FieldOptions) (err error) {
+func (cdc *Codec) decodeReflectJSONStruct(bz []byte, info *TypeInfo, rv reflect.Value, _ FieldOptions) (err error) {
 	if !rv.CanAddr() {
 		panic("rv not addressable")
 	}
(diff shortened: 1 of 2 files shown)


json-decode.go
```diff
@@ -362,7 +362,7 @@ func (cdc *Codec) decodeReflectJSONSlice(bz []byte, info *TypeInfo, rv reflect.V
 }
 
 // CONTRACT: rv.CanAddr() is true.
-func (cdc *Codec) decodeReflectJSONStruct(bz []byte, info *TypeInfo, rv reflect.Value, fopts FieldOptions) (err error) {
+func (cdc *Codec) decodeReflectJSONStruct(bz []byte, info *TypeInfo, rv reflect.Value, _ FieldOptions) (err error) {
```
melekes: is it possible to remove the last parameter? don't see a reason to keep it

melekes (commented):
👍

json-decode.go
NicolasMahe: yes it's possible. i didn't remove it because the binary version has it:
https://github.com/tendermint/go-amino/blob/59d50ef176f686dc2c0c150e85a9381a78b9978a/binary-decode.go#L763-L764

NicolasMahe: updated 👍

melekes (approved):

jacquesvcritien:
Are there plans for this to be merged please?

melekes:
> Are there plans for this to be merged please?

I don't think this repo is being maintained. Better to fork and continue from there.

PR #316, opened 2020-05-12
Add codec method to retrieve a Go type registered name

gsora:
Given a codec, users might want to know what's the Amino type associated
with a given struct instance.

This commit adds the functionality needed to do that, for both instance and pointer-to-instance
types.

Nil interfaces are rejected by default.

Method completed with tests.

---

We are in the process of writing various SDKs for our project ([commercionetwork](http://github.com/commercionetwork/commercionetwork)), and for the Go one we thought it might be a great idea to simply reuse all the types implemented in our modules.

We need to know at runtime Amino message types for every possible type in order to construct a well-defined Cosmos REST transaction message.

Right now, we manually parse `codec.PrintTypes()` output and build a map which associates Go type to its Amino name.

This patch integrates basically the same idea, by leveraging the codec's integrated type map.

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -308,6 +308,30 @@ func (cdc *Codec) PrintTypes(out io.Writer) error {
 	return nil
 }
 
+// ConcreteRegisteredName returns the name under which i has been registered.
+func (cdc *Codec) ConcreteRegisteredName(i interface{}) (string, error) {
+	cdc.mtx.RLock()
+	defer cdc.mtx.RUnlock()
+
+	if i == nil {
+		return "", errors.New("argument must not be nil")
+	}
+
+	gt := reflect.TypeOf(i)
+
+	// if i is a non-nil pointer, dereference it and grab its inner Elem
+	if gt.Kind() == reflect.Ptr {
+		gt = gt.Elem()
+	}
+
+	ct, ok := cdc.typeInfos[gt]
+	if !ok {
+		return "", fmt.Errorf("no type registered for %s", gt.String())
+	}
+
+	return ct.Name, nil
+}
+
 // A heuristic to guess the size of a registered type and return it as a string.
 // If the size is not fixed it returns "variable".
 func getLengthStr(info *TypeInfo) string {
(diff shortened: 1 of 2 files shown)


codec.go
```diff
@@ -308,6 +308,30 @@ func (cdc *Codec) PrintTypes(out io.Writer) error {
...
+
+	if i == nil {
+		return "", errors.New("argument must not be nil")
+	}
+
+	gt := reflect.TypeOf(i)
+
+	// if i is a non-nil pointer, dereference it and grab its inner Elem
+	if gt.Kind() == reflect.Ptr {
+		gt = gt.Elem()
```
liamsi, line 324: What happens if this a **SomeType? (pointer to pointer)

codec.go
```diff
@@ -308,6 +308,30 @@ func (cdc *Codec) PrintTypes(out io.Writer) error {
...
+	}
+
+	ct, ok := cdc.typeInfos[gt]
+	if !ok {
+		return "", fmt.Errorf("no type registered for %s", gt.String())
```
liamsi, line 329: This should probably be a sentinel error, such that callers can switch over a returned error to see if the type wasn't registered or if any other error occurred.

And independent of that: I think `%s` already calls `String()` on the type (no need to explicitly call it here).

liamsi (commented):
Looks good to me. Usually users search for the types and codecs in the relevant code base currently (e.g. tendermint/cosmos-sdk). Maybe it's cool to have a simple way look them up instead 👍

codec.go
gsora: Uhm, **SomeType would return error even though the struct type has been registered previously.

I could add a check on the inner gt.Elem(), this way it would dereference 2 times and get to the concrete instance.

Thoughts?

---

Something like:

```go
gt := reflect.TypeOf(i)

for gt.Kind() == reflect.Ptr {
	gt = gt.Elem()
}
```

codec.go
gsora: Would those two errors be okay for the purpose?
 - `ErrNilPointer`, when `i` is `nil`;
 - `ErrTypeNotRegistered`, when the concrete type of `i` is not registered;

Yup, `String()` will be removed :-)

PR #323, opened 2020-09-10
Add support for (Un)MarshalAminoJSON override

amaury1093:
ref: https://github.com/cosmos/cosmos-sdk/issues/7109#issuecomment-689709943

If a type implements `MarshalAminoJSON` and `UnmarshalAminoJSON`, then amino will use these methods for JSON marshaling.

This is the json-equivalent for methods `MarshalAmino` and `UnmarshalAmino`

also cc @alexanderbez, since I can't add you as a reviewer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## [Unreleased]
 
+## 0.16.0 (September 11, 2020)
+
+IMPROVEMENTS:
+ - Add support for `Un/MarshalAminoJSON` override: if a type implements
+ `Un/MarshalAminoJSON`, then amino will use these methods for JSON un/marshalling
+ ([#323]).
+
+[#323]: https://github.com/tendermint/go-amino/pull/323
+
 ## 0.15.1 (October 10, 2019)
 
 ### IMPROVEMENTS:
--- a/codec.go
+++ b/codec.go
@@ -88,10 +88,10 @@ type ConcreteInfo struct {
 	AminoMarshalReprType       reflect.Type // <ReprType>
 	IsAminoUnmarshaler         bool         // Implements UnmarshalAmino(<ReprObject>) (error).
 	AminoUnmarshalReprType     reflect.Type // <ReprType>
-	IsAminoMarshalerJSON       bool         // Implements MarshalAminoJSON() (<ReprObject>, error).
-	AminoMarshalJSONReprType   reflect.Type // <ReprType>
-	IsAminoUnmarshalerJSON     bool         // Implements UnmarshalAminoJSON(<ReprObject>) (error).
-	AminoUnmarshalJSONReprType reflect.Type // <ReprType>
+	IsAminoJSONMarshaler       bool         // Implements MarshalAminoJSON() (<ReprObject>, error).
+	AminoJSONMarshalReprType   reflect.Type // <ReprType>
+	IsAminoJSONUnmarshaler     bool         // Implements UnmarshalAminoJSON(<ReprObject>) (error).
+	AminoJSONUnmarshalReprType reflect.Type // <ReprType>
 }
 
 type StructInfo struct {
--- a/codec.go
+++ b/codec.go
@@ -84,10 +84,14 @@ type ConcreteInfo struct {
 
 	// These fields get set for all concrete types,
 	// even those not manually registered (e.g. are never interface values).
-	IsAminoMarshaler       bool         // Implements MarshalAmino() (<ReprObject>, error).
-	AminoMarshalReprType   reflect.Type // <ReprType>
-	IsAminoUnmarshaler     bool         // Implements UnmarshalAmino(<ReprObject>) (error).
-	AminoUnmarshalReprType reflect.Type // <ReprType>
+	IsAminoMarshaler           bool         // Implements MarshalAmino() (<ReprObject>, error).
+	AminoMarshalReprType       reflect.Type // <ReprType>
+	IsAminoUnmarshaler         bool         // Implements UnmarshalAmino(<ReprObject>) (error).
+	AminoUnmarshalReprType     reflect.Type // <ReprType>
+	IsAminoMarshalerJSON       bool         // Implements MarshalAminoJSON() (<ReprObject>, error).
+	AminoMarshalJSONReprType   reflect.Type // <ReprType>
+	IsAminoUnmarshalerJSON     bool         // Implements UnmarshalAminoJSON(<ReprObject>) (error).
+	AminoUnmarshalJSONReprType reflect.Type // <ReprType>
 }
 
 type StructInfo struct {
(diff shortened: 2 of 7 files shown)


alessio:
> also cc @alexanderbez, since I can't add you as a reviewer

Added. And you should be in the position to add him in future PRs as well :+1:

repr_test.go
```diff
@@ -97,3 +97,54 @@ func TestMarshalAminoJSON(t *testing.T) {
...
+
+type Bar struct {
+	a string
+	b int
+	c []*Bar
+	D string // exposed
+}
+
+func (b Bar) MarshalAminoJSON() ([]pair, error) { // nolint: golint
+	return []pair{
```
alexanderbez, line 108: Curious, why does this return a list of pairs and not bytes? What is the typical implementation of `MarshalAminoJSON` look like?

amaury1093: Yeah, it can return whatever it wants. I just copied the `MarshalAmino` tests, which returned pairs. Can change to []byte, if you prefer that.

In the SDK we'll want to return []byte, like here: https://github.com/cosmos/cosmos-sdk/pull/7276/files#diff-1f6fdece2a5550f27ed0b77247d58f8bR60-R65

alexanderbez (approved):

json-encode.go
```diff
@@ -61,6 +82,17 @@ func (cdc *Codec) encodeReflectJSON(w io.Writer, info *TypeInfo, rv reflect.Valu
...
+		if rv.Addr().Type().Implements(jsonMarshalerType) {
+			err = invokeMarshalJSON(w, rv.Addr())
+			return
+		}
+	} else if rv.Type().Implements(jsonMarshalerType) {
+		err = invokeMarshalJSON(w, rv)
+		return
+	}
+
+	// Handle override if rv implements MarshalAmino.
```
aaronc: Why is this needed?

codec.go
```diff
@@ -86,10 +86,14 @@ type ConcreteInfo struct {
...
-	IsAminoUnmarshaler     bool         // Implements UnmarshalAmino(<ReprObject>) (error).
-	AminoUnmarshalReprType reflect.Type // <ReprType>
+	IsAminoMarshaler           bool         // Implements MarshalAmino() (<ReprObject>, error).
+	AminoMarshalReprType       reflect.Type // <ReprType>
+	IsAminoUnmarshaler         bool         // Implements UnmarshalAmino(<ReprObject>) (error).
+	AminoUnmarshalReprType     reflect.Type // <ReprType>
+	IsAminoMarshalerJSON       bool         // Implements MarshalAminoJSON() (<ReprObject>, error).
+	AminoMarshalJSONReprType   reflect.Type // <ReprType>
+	IsAminoUnmarshalerJSON     bool         // Implements UnmarshalAminoJSON(<ReprObject>) (error).
+	AminoUnmarshalJSONReprType reflect.Type // <ReprType>
```
aaronc: nit
```suggestion
	IsAminoJSONMarshaler       bool         // Implements MarshalAminoJSON() (<ReprObject>, error).
	AminoJSONMarshalReprType   reflect.Type // <ReprType>
	IsAminoJSONUnmarshaler     bool         // Implements UnmarshalAminoJSON(<ReprObject>) (error).
	AminoJSONUnmarshalReprType reflect.Type // <ReprType>
```

json-decode.go
```diff
@@ -61,6 +61,29 @@ func (cdc *Codec) decodeReflectJSON(bz []byte, info *TypeInfo, rv reflect.Value,
...
+		}
+		// Then, decode from repr instance.
+		uwrm := rv.Addr().MethodByName("UnmarshalAminoJSON")
+		uwouts := uwrm.Call([]reflect.Value{rrv})
+		erri := uwouts[0].Interface()
+		if erri != nil {
+			err = erri.(error)
+		}
+		return
+	}
```
aaronc: Why can't we just do:
```go
	// Handle override if a pointer to rv implements json.Unmarshaler.
	if rv.Addr().Type().Implements(jsonUnmarshalerType) {
		err = rv.Addr().Interface().(AminoJSONUnmarshaler).UnmarshalAminoJSON(bz)
		return
	}
```

Is this because we want to allow any type to be used instead of `[]byte`? Is it better to be consistent with `Un/MarshalAmino` or should we just do the simpler thing and use the same signatures as `Un/MarshalJSON`?

json-encode.go
```diff
@@ -49,6 +49,27 @@ func (cdc *Codec) encodeReflectJSON(w io.Writer, info *TypeInfo, rv reflect.Valu
...
+			return
+		}
+		rinfo, err = cdc.getTypeInfoWlock(info.AminoMarshalJSONReprType)
+		if err != nil {
+			return
+		}
+		// Then, encode the repr instance.
+		err = cdc.encodeReflectJSON(w, rinfo, rrv, fopts)
+		return
+	}
```
aaronc: ditto

aaronc: Ditto, can't we just cast the type and call `MarshalAminoJSON`?

json-encode.go
```diff
@@ -61,6 +82,17 @@ func (cdc *Codec) encodeReflectJSON(w io.Writer, info *TypeInfo, rv reflect.Valu
...
 	// Handle override if rv implements json.Marshaler.
+	if rv.CanAddr() { // Try pointer first.
+		if rv.Addr().Type().Implements(jsonMarshalerType) {
+			err = invokeMarshalJSON(w, rv.Addr())
+			return
+		}
+	} else if rv.Type().Implements(jsonMarshalerType) {
+		err = invokeMarshalJSON(w, rv)
+		return
+	}
```
amaury1093: note to self: to be removed

json-decode.go
amaury1093: > Is it better to be consistent with Un/MarshalAmino or should we just do the simpler thing and use the same signatures as Un/MarshalJSON

Yes, that's the whole debate. The way I went for is to keep the symmetry in function signature between:
- Un/MarshalJSON and Un/MarshalBinary
- Un/MarshalAminoJSON and Un/MarshalAmino

json-encode.go
amaury1093: removed, a stray copy-paste.

aaronc (approved):
utACK

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## 0.15.1 (September 10, 2020)
```
aaronc: `0.15.1` is already tagged on a different branch: https://github.com/tendermint/go-amino/blob/release/v0.15.1/CHANGELOG.md. I think we should put these changes in `0.16.0`. It appears that there are also some changes in `0.15.1` that haven't been merged back into master. @marbar3778 you did the last point release. Are you able to give us some guidance on doing a new release here?

aaronc (changes requested):

CHANGELOG.md
tac0turtle: I believe the best approach would be to branch off of the most recent release branch or tag. It is unclear if master is working correctly, I vaguely remember there being a bug or two on master that were never fixed.

amaury1093: Done. There's a changelog entry indeed for 0.15.1, but only in this release branch. I added a 0.16.0 one.

tac0turtle: Sorry for not being descriptive. Could you create a branch from release/0.15.1 like rc0/0.16 then put this PR against that?

amaury1093: Ah ok, understood.

aaronc (approved):

